PR #9, opened 2015-03-03
Update addressof.hpp

akumta:
To update ticket# 10311

### DIFF
--- a/include/boost/core/addressof.hpp
+++ b/include/boost/core/addressof.hpp
@@ -107,7 +107,7 @@ template<class T>
 BOOST_FORCEINLINE
 T * addressof( T & v )
 {
-#if (defined( __BORLANDC__ ) && BOOST_WORKAROUND( __BORLANDC__, BOOST_TESTED_AT( 0x610 ) ) ) || defined( __SUNPRO_CC )
+#if (defined( __BORLANDC__ ) && BOOST_WORKAROUND( __BORLANDC__, BOOST_TESTED_AT( 0x610 ) ) ) || (defined(__SUNPRO_CC) && (__SUNPRO_CC <= 0x5120))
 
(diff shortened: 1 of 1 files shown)


pdimov:
Wouldn't it be better if we use BOOST_WORKAROUND for consistency?

Lastique:
Frankly, I always found BOOST_WORKAROUND/BOOST_TESTED_AT confusing.

pdimov:
In this case it would not use BOOST_TESTED_AT. `defined(__SUNPRO_CC) && BOOST_WORKAROUND(__SUNPRO_CC, <= 0x5120)`.

The idea of BOOST_WORKAROUND is both to mark something as a workaround and to make it possible for all workarounds to be turned off (so that the supposedly conforming code is chosen.)

In principle, the original should already have used BOOST_WORKAROUND, this time with BOOST_TESTED_AT, like `defined(__SUNPRO_CC) && BOOST_WORKAROUND(__SUNPRO_CC, BOOST_TESTED_AT( 0x590))`. This says that all known versions need the workaround and the latest version tested was 5.9. When 5.13 is then discovered to no longer need the workaround, we change BOOST_TESTED_AT with the appropriate <= test.

Lastique:
Thanks for the explanation. The problem is when I read the code I don't immediately know what `BOOST_WORKAROUND(__SUNPRO_CC, BOOST_TESTED_AT(0x590))` means. Is this condition true only for 0x590 version? What about other versions, are they not tested with the workaround? What in fact that means is 'always return `true` unless `BOOST_DETECT_OUTDATED_WORKAROUNDS` is defined'. So the version number doesn't really matter. This gives me a stutter.

The sole `BOOST_WORKAROUND` is better in this respect because I can at least see the condition.

pdimov:
Yes, you have to know what it means. But once you know, it's not really possible to misinterpret.

PR #11, opened 2015-06-05
Disable auto_ptr support in get_pointer when BOOST_NO_AUTO_PTR is def…

awulkiew:
…ined.

This is consistent with how Boost.SmartPtr handles deprecated std::auto_ptr. This way the user can disable the auto_ptr support to suppress compiler warnings.


pdimov:
Thanks Adam.

PR #16, opened 2015-08-13
Enable_if_has_type

jfalcou:
Provides an autonomous of enable_if_has_type which enable SFIANE checks on type existence.
Renamed enable_if_has_type after discussions on the Bosot dev ML.


jfalcou:
crap yes, it fell through :/

PR #19, opened 2015-12-31
Better detection of cxxabi header for gcc+libc++

apolukhin:
Some of the Android related tests run GCC with libc++. On those tests abi::__cxa_demangle is not used, while it is available.
This pull request allows usage of `__has_include` for latest GCC and assumes that all the libc++ have `cxxabi.h` header.

### DIFF
--- a/include/boost/core/demangle.hpp
+++ b/include/boost/core/demangle.hpp
@@ -20,7 +20,7 @@
 // __has_include is currently supported by GCC and Clang. However GCC4.9 may have issues and
 // returns 1 for `defined( __has_include )`, while `__has_include` is actually not supported:
 // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=63662
-#if (!defined( __GNUC__ ) || __GNUC__ >= 5) && defined( __has_include )
+#if (defined( __clang__ ) || !defined( __GNUC__ ) || __GNUC__ >= 5) && defined( __has_include )
 # if __has_include(<cxxabi.h>)
 #  define BOOST_CORE_HAS_CXXABI_H
 # endif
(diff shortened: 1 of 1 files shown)


Lastique:
1. Your change seems to drop support for clang as it defines `__GNUC__` but not to 5 (to 4).
2. I'm not sure the presence of libc++ indicates presence of `cxxabi.h`. I'd rather be more conservative and accidentally not perform demangling instead of failing to compile because of the absent header.

apolukhin:
> 1. Your change seems to drop support for clang as it defines __GNUC__ but not to 5 (to 4).

Fixed that.

> 2. I'm not sure the presence of libc++ indicates presence of cxxabi.h. I'd rather be more conservative and accidentally not perform demangling instead of failing to compile because of the absent header.

Search for "cxxabi.h libc++" gave no pages that describe relevant bugs. I'd rather commit it in that way and remove the `|| defined( _LIBCPP_VERSION )` in case of any errors in Boost test suite.

Lastique:
You didn't find anything because apparently noone uses libc++ macros as an indication of `cxxabi.h` presence. This makes sense since, AFAIU, `cxxabi.h` is part of gcc runtime and, if anything, can be considered part of libstdc++. libc++ has no relation to it, an I suspect platforms that run libc++ but don't have `cxxabi.h` are quite real (OS X comes to mind).

So I committed a different change that makes use of `__has_include` on gcc, when possible, but otherwise only use `cxxabi.h` when using libstdc++. If you can guarantee that a particular platform uses libc++ and always have `cxxabi.h` as well then please create a separate PR adding detection of that particular platform (e.g. checking for OS and compiler).

PR #20, opened 2016-05-19
Added BOOST_TEST_NOT to lightweight test.

eldiener:


pdimov:
What is the purpose of this change?

eldiener:
Purely ease of use. Manually specifying BOOST_TEST(!(expr)) is a pain. It also parallels the BOOST_TEST_EQ/BOOST_TEST_NE pair in its own way.

eldiener:
It is a convenient way for testing that some expression is not true.

pdimov:
I'd like to see an example where it's an improvement over the equivalent BOOST_TEST, if you don't mind.

And a doc patch. :-)

eldiener:
Is there documentation for lightweight_test ? I have never found any.

pdimov:
There wasn't, but there is now, in `doc/lightweight_test.qbk`.

eldiener:
A simple example might be:

`
#include <boost/type_traits.hpp>
#include <boost/detail/lightweight_test.hpp>

int main()
    {
    BOOST_TEST_NOT(boost::is_compound<int>::value);
    return boost::report_errors();
    }
`

I will update the doc and tests accordingly as part of the pull request.

pdimov:
That's `BOOST_TEST_TRAIT_FALSE((boost::is_compound<int>))`.

eldiener:
That's nice to know. thanks ! But my point is still that BOOST_TEST_NOT is still a syntactical convenience in any situation whenever you want to test that some expression should not be true.

Lastique:
Frankly, I don't see much advantage compared to prepending '!' to the expression (or changing comparison operator to the opposite). Just my 2 cents.

eldiener:
Do you want me to close this PR. It seems you have no intention of accepting it.

Lastique:
I'm leaving that to Peter to decide as he's the main maintainer of that header.

pdimov:
Merged, sorry for the delay and thanks for the patience.

PR #22, opened 2017-01-30
Implement constexpr addressof

glenfe:
Putting the pull request up for discussion: C++17 provides constexpr addressof, which we can provide in C++11 and above when specific language features are supported by the implementation (e.g. expression SFINAE and constexpr). Do we want this?

I will add the MSVC workarounds back and update the pull request. The Sun and Borland workarounds don't have to be removed either.

### DIFF
--- a/include/boost/core/addressof.hpp
+++ b/include/boost/core/addressof.hpp
@@ -1,162 +1,250 @@
-// Copyright (C) 2002 Brad King (brad.king@kitware.com)
-//                    Douglas Gregor (gregod@cs.rpi.edu)
-//
-// Copyright (C) 2002, 2008, 2013 Peter Dimov
-//
-// Distributed under the Boost Software License, Version 1.0. (See
-// accompanying file LICENSE_1_0.txt or copy at
-// http://www.boost.org/LICENSE_1_0.txt)
+/*
+Copyright (C) 2002 Brad King (brad.king@kitware.com)
+                   Douglas Gregor (gregod@cs.rpi.edu)
 
-// For more information, see http://www.boost.org
+Copyright (C) 2002, 2008, 2013 Peter Dimov
+
+Copyright (C) 2017 Glen Joseph Fernandes (glenjofe@gmail.com)
+
+Distributed under the Boost Software License, Version 1.0. (See
+accompanying file LICENSE_1_0.txt or copy at
+http://www.boost.org/LICENSE_1_0.txt)
+*/
 
 #ifndef BOOST_CORE_ADDRESSOF_HPP
 #define BOOST_CORE_ADDRESSOF_HPP
 
-# include <boost/config.hpp>
-# include <boost/detail/workaround.hpp>
-# include <cstddef>
+#include <boost/config.hpp>
+#include <boost/detail/workaround.hpp>
+#include <cstddef>
 
-namespace boost
-{
+#if defined(BOOST_NO_SFINAE_EXPR) || \
+    defined(BOOST_NO_CXX11_RVALUE_REFERENCES) || \
+    defined(BOOST_NO_CXX11_CONSTEXPR) || \
+    defined(BOOST_NO_CXX11_DECLTYPE) || \
+    (defined(BOOST_MSVC) && \
+        BOOST_WORKAROUND(BOOST_MSVC, BOOST_TESTED_AT(1900)))
+#define BOOST_CORE_NO_CONSTEXPR_ADDRESSOF
+#endif
 
-namespace detail
-{
+namespace boost {
+namespace core {
+namespace detail {
 
-template<class T> struct addr_impl_ref
-{
-    T & v_;
-
-    BOOST_FORCEINLINE addr_impl_ref( T & v ): v_( v ) {}
-    BOOST_FORCEINLINE operator T& () const { return v_; }
-
+template<class T>
+class addressof_ref {
+public:
+    BOOST_FORCEINLINE addressof_ref(T& o) BOOST_NOEXCEPT
+        : o_(o) { }
+    BOOST_FORCEINLINE operator T&() const BOOST_NOEXCEPT {
+        return o_;
+    }
 private:
-    addr_impl_ref & operator=(const addr_impl_ref &);
+    T& o_;
 };
 
-template<class T> struct addressof_impl
-{
-    static BOOST_FORCEINLINE T * f( T & v, long )
-    {
-        return reinterpret_cast<T*>(
-            &const_cast<char&>(reinterpret_cast<const volatile char &>(v)));
+template<class T>
+struct address_of {
+    static BOOST_FORCEINLINE T* get(T& o, long) BOOST_NOEXCEPT {
+        return reinterpret_cast<T*>(&
+            const_cast<char&>(reinterpret_cast<const volatile char&>(o)));
     }
-
-    static BOOST_FORCEINLINE T * f( T * v, int )
-    {
-        return v;
+    static BOOST_FORCEINLINE T* get(T* p, int) BOOST_NOEXCEPT {
+        return p;
     }
 };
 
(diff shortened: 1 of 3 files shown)


Lastique:
> Do we want this?

Why not? Provided that compatibility with C++03 is preserved, I think `boost::addressof` should provide the same services as `std::addressof`.

> I will add the MSVC workarounds back and update the pull request. The Sun and Borland workarounds don't have to be removed either.

Also, `BOOST_FORCEINLINE` don't have to be removed.

pdimov:
I may come across as overly negative, but this sounds like lot of potential problems for zero upside.

glenfe:
The use case that comes to mind is a conforming C++17 optional implementation which requires constexpr addressof (which is why libstdc++, libc++, Andrezj et al implemented much of the above).

Two alternatives could be to provide it with a different name, e.g. constexpr_addressof, or require (via macro) explicitly choosing the above functionality.

Lastique:
On 01/30/17 16:29, Glen Fernandes wrote:
> The use case that comes to mind is a conforming C++17 optional
> implementation which requires constexpr addressof (which is why
> libstdc++, libc++, Andrezj et al implemented much of the above).
>
> Two alternatives could be to provide it with a different name, e.g.
> constexpr_addressof, or require (via macro) explicitly choosing the
> above functionality.

Why the different naming, let alone config macro? Is there a problem 
with implementing the universal `addressof`?

glenfe:
Not that I know of (I was deferring to Peter's instincts for predicting trouble). I've used the above implementation (or close to it) for our implementation of optional without any issues for a while.

Lastique:
On 01/30/17 17:25, Glen Fernandes wrote:
> Not that I know of (I was deferring to Peter's instincts for predicting
> trouble). I've used the above implementation (or close to it) for our
> implementation of optional without any issues for a while.

I think the problems, if they appear, can be solved. That should not 
stop us from moving forward with `addressof`.

pdimov:
I updated .travis.yml to test on more compilers and more modes (C++03/11/14) as the current one was only a quick C++03 smoke check. This was fine for what we had but wasn't enough to test this change. The test time jumped from a few minutes to a few hours but what could one do.

Glen, would you please merge the .travis.yml change and push the branch again to kick off Travis?

apolukhin:
You could also use the feature testing macro `__cpp_lib_array_constexpr` to detect `constexpr std::addressof` availability ([p0031r0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0031r0.html), that adds the feature testing macro, can not be implemented without `constexpr std::addressof`).

So if `__cpp_lib_array_constexpr` defined - just use `std::addressof` inside the `boost::addressof`.

pdimov:
If we already have a working boost::addressof implementation, it's not clear to me why we'd need to use std::addressof instead (when it's present and constexpr).

glenfe:
@pdimov Will do.

K-ballo:
>  Two alternatives could be to provide it with a different name, e.g.
>  constexpr_addressof, or require (via macro) explicitly choosing the
>  above functionality.
> Why the different naming, let alone config macro? Is there a problem
> with implementing the universal `addressof`?

The universal `addressof` cannot be implemented in plain C++, it requires a builtin.

glenfe:
The implementations are starting to provide builtins, that Agustín pointed out. We should use them for addressof (once we've verified they pass all tests). For older compilers without the C++17 builtin, the question still remains as to whether to provide the emulated functionality as addressof or under a new name (e.g. constexpr_addressof).

One advantage of the latter could be as simple (but no less important) as not burdening the users with the compile-time overhead of the expression SFINAE detection, which I realize isn't an unreasonable desire.

glenfe:
A note about the `defined(BOOST_MSVC)` check: It is because `BOOST_NO_SFINAE_EXPR` is not defined for VC14 Update 3. If VC15 eventually supports these cases by improving its SFINAE support, it can be relaxed to a `(BOOST_MSVC < 1910)` check.

glenfe:
Builds passed with all compilers. Tomorrow I'll update the pull request with usage of `__builtin_addressof` on implementations which support it.

pdimov:
Looks good to me.

Re `defined(BOOST_MSVC)`, it should be `defined(BOOST_MSVC) && BOOST_WORKAROUND(BOOST_MSVC, BOOST_TESTED_AT(1900))`. For stylistic reasons you could separate it in its own `#if` block and add a comment. (The idea of using `BOOST_WORKAROUND` is to allow compiler vendors to switch the workarounds off and see if their new version, VC15 in this case, works.)

Similarly, the `defined(__INTEL_COMPILER)` workaround should also use `BOOST_WORKAROUND`.

I'm a bit concerned about using generic names such as `make_void` here but don't see an immediate problem with that.

You can merge when ready as far as I'm concerned.

Lastique:
gcc 4.7 failed some tests after this change:

`../boost/core/addressof.hpp:235:44: error: 'operator&' not defined`

http://www.boost.org/development/tests/develop/developer/output/teeks99-02-dg4-7-Docker-64on64-boost-bin-v2-libs-atomic-test-atomicity-test-gcc-4-7~c++11-debug-threading-multi.html

http://www.boost.org/development/tests/develop/developer/output/Sandia-gcc-4-7-2-c++11-boost-bin-v2-libs-atomic-test-atomicity-test-gcc-4-7-2-debug-threading-multi.html

pdimov:
Looks like our `reference_wrapper` tests don't test with `volatile`. I'll add a quick test for that.

Also, I'm getting lots of

```
1>c:\Projects\boost-git\boost\boost/core/addressof.hpp(44) : warning C4512: 'boost::core::detail::addressof_ref<T>' : assignment operator could not be generated
```

from MSVC.

glenfe:
Looks like gcc 4.8+ fixed those issues; I'll exclude 4.7. I had removed the `operator=` declaration because I didn't think it was necessary; I'll add it back to suppress the warnings.

pdimov:
Hmm, `addressof_test` already tests with `volatile`, why isn't this triggering the error on 4.7? Odd.

glenfe:
Actually, we can support 4.7 too if we enable the Intel workaround for it.  i.e. The presence of the following:
```
#if BOOST_WORKAROUND(BOOST_INTEL, < 1600)
struct addressof_addressable { };
addressof_addressable* operator&(addressof_addressable&) BOOST_NOEXCEPT;
#endif
```

The condition just needs to be:
`BOOST_WORKAROUND(BOOST_INTEL, < 1600) || BOOST_WORKAROUND(BOOST_GCC, < 40800)`

glenfe:
Also, the `#if defined(macro) && BOOST_WORKAROUND(macro, check)` can just be expressed more succinctly as `#if BOOST_WORKAROUND(macro, check)`, right? (Is there any reason to prefer the former?)

pdimov:
The usual - warning suppression when the macro isn't defined (`-Wundef`).

pdimov:
Although going by https://svn.boost.org/trac/boost/ticket/1130 and the referenced https://svn.boost.org/trac/boost/changeset/39473/trunk/boost/detail/workaround.hpp, `BOOST_MSVC` and `BOOST_INTEL` are special-cased to not warn, so maybe I'm behind the times on this one. :-)

pdimov:
Nine years behind, to be specific.

pdimov:
> Hmm,  `addressof_test` already tests with `volatile`, why isn't this triggering the error on 4.7? Odd.

Because the Travis matrix didn't have g++ 4.7 in C++11 mode. Mystery solved, I guess. Added.

pdimov:
Still doesn't work with a the private `operator&` member.

glenfe:
Affirmative. If I don't find a 4.7 workaround tonight, I'll exclude it and limit the functionality to 4.8 and higher.

PR #27, opened 2017-02-18
Added BOOST_TEST_ALL_WITH macro to compare container contents with a …

breese:
…predicate


pdimov:
This looks good. Could you however split the test into two, pass/fail? The current arrangement makes it possible for one of the pass test cases to fail, one of the fail test cases to pass, and the overall count to remain correct.

PR #28, opened 2017-04-16
Document an emulation limitation of scoped enum

Sawyer47:
I recently stumbled upon a case in which scoped enums have different behaviors in C++03 and C++11. I thought it would be a good idea to document this case.

### DIFF
--- a/doc/scoped_enum.qbk
+++ b/doc/scoped_enum.qbk
@@ -67,9 +67,9 @@ The enumeration can be forward declared:
 
     BOOST_SCOPED_ENUM_FORWARD_DECLARE(future_errc);
 
-There are however some limitations. First, the emulated scoped enum is not a C++ enum, so `is_enum< future_errc >` will be `false_type`.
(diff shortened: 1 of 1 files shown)


Lastique:
I wonder if we should consider this a bug instead.

viboes:
Hi and thanks for reporting.

I agree that it is not clear from the documentation that the type of `enum_scoped::a` is not `enum_scoped`, but `enum_scoped::type` which of course is an enumeration. 

Your patch is welcome, and I suggest it is merged as soon as boost 1.64 is delivered.

It could be possible to define constants `a` and `b` of type `enum_scoped` (with a different macro syntax), but these constants couldn't be used in the switch/case. 
I don't know how to have enumerators and constants with the same name. 
I don't know which names should be used for them; a for the enumerator and a_ for the class instance? 

I agree the workaround `enum_scoped(enum_scoped::a)` is cumbersome, but it works.  

I've opted for this partial solution as the more convenient.

I don't think this is a bug we can fix without adding other macros.

HDembinski:
Hi, just passing by and wondering why the PR was not accepted after all?

pdimov:
Good question. Any objections to merging it?

PR #30, opened 2017-04-23
Also disable addressof for r-values when possible

glenfe:
Now in C++17: `template<class T> const T* addressof(const T&&) = delete;`


Lastique:
Doesn't this disable `addressof` for all non-const types? Shouldn't there be `disable_if_c< is_reference< T >::value`?

pdimov:
`T` can never be a reference here.

Lastique:
Ah, I forgot there is already an overload for lvalue reference. Sorry for the noise.

PR #33, opened 2017-05-16
Add pointer_traits with to_address

glenfe:
Still working on documentation.

### DIFF
--- a/include/boost/core/pointer_traits.hpp
+++ b/include/boost/core/pointer_traits.hpp
@@ -1,0 +1,256 @@
+/*
+Copyright 2017 Glen Joseph Fernandes
+(glenjofe@gmail.com)
+
+Distributed under the Boost Software License, Version 1.0.
+(http://www.boost.org/LICENSE_1_0.txt)
+*/
+#ifndef BOOST_CORE_POINTER_TRAITS_HPP
+#define BOOST_CORE_POINTER_TRAITS_HPP
+
+#include <boost/config.hpp>
+#include <boost/detail/workaround.hpp>
+#if !defined(BOOST_NO_CXX11_ALLOCATOR) && \
+    !BOOST_WORKAROUND(BOOST_MSVC, < 1910)
+#include <memory>
+#else
+#include <boost/core/addressof.hpp>
+#endif
+
+namespace boost {
+
+template<class T>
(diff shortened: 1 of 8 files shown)


pdimov:
So you decided to call it `to_raw_pointer` after all. Do we need it now that the implementation of `pointer_traits` doesn't use it?

`boost::pointer_traits` should be present even if `std::pointer_traits` isn't, and should work the same way, except of course `rebind` if aliases are unavailable. We should have a convention for that... `rebind_trait`? `rebind_type`?

glenfe:
I'm on the fence between: Keeping it at all (with any name, `to_raw_pointer`, `to_address`, `get_address`) or dropping it. (`pointer_to_address` isn't too long, but the convention of `to_string` is nice, one less type mentioned in identifier).

I wish `rebind` were specified as types containing a `type` or `other` member instead of aliases, just for compatibility sake. If `std::pointer_traits` isn't available, should we just provide the `T*` specialization of `pointer_traits`, along with a `rebind` containing `other`?

glenfe:
I've found myself writing something like the following more than once: https://github.com/boostorg/smart_ptr/blob/develop/include/boost/smart_ptr/allocate_shared_array.hpp#L134-L144

Were you thinking of something like that, but where the traits template is also specified (i.e. pointer_traits, allocator_traits)?

pdimov:
> If std::pointer_traits isn't available, should we just provide the T* specialization of pointer_traits, along with a rebind containing other?

I was thinking that we should provide a fully functional `pointer_traits` if the standard one isn't available, but with `rebind` replaced with a struct with some other name (`rebind_type`?), to not conflict with the standard one. We can supply that other name in `boost::pointer_traits` in the other path as well, so that code can use it without ifdefs.

Incidentally, as written `boost::pointer_traits<>::rebind` returns `std::pointer_traits<>`, so we'll have to massage that in either case.

pdimov:
`template<class U> using rebind = mp_rename<std::pointer_traits<T>::rebind<U>, boost::pointer_traits>;` :-)

pdimov:
Ah wait, this `rebind` returns the pointer type. Never mind.

glenfe:
Merged after Peter's review and approval on Slack.

PR #37, opened 2017-12-25
address_of.hpp: Fix namespace ambiguity

jzmaddock:
I'm seeing compiler errors with vc12, when testing serialization of Boost.Multiprecision types - there is ambiguity between boost::detail and boost::multiprecision::detail.  I assume the error occurs for serialization of any type that has it's own detail namespace.  The fix is trivially to qualify the use of "detail::".


glenfe:
Merged.

Issue #38, opened 2018-08-03
Add tests for floats with lightweight_tests


HDembinski:
It would be nice to have a test for floats. Something like this:
```
#define BOOST_TEST_IS_CLOSE(a, b, eps) BOOST_TEST(std::abs(a - b) < eps)
```
I am happy to make a PR if that helps.

breese:
I second that request.

I have created one as an extension for my own projects. It uses a separate implementation function in order to print the argument of failed tests. The tolerance can be specified.

My implementation looks like this (modulo the test macro)
```
template <typename T>
void test_close_impl(char const * expr1,
                     char const * expr2,
                     char const * file, int line, char const * function,
                     const T& lhs, const T& rhs, T tolerance)
{
    if (std::fabs(lhs - rhs) <= tolerance)
    {
        boost::detail::report_errors_remind();
    }
    else
    {
        BOOST_LIGHTWEIGHT_TEST_OSTREAM
            << file << "(" << line << "): test '" << expr1 << " == " << expr2
            << "' failed in function '" << function << "': "
            << "'" << lhs << "' != '" << rhs << "'" << std::endl;
        ++boost::detail::test_errors();
    }
}
```

HDembinski:
Makes sense to generate a custom message for this case. Maybe even note the tolerance in the message? Something like x != y (tolerance: z).

breese:
Just to float (pun intended) another idea.

We could define a `with_tolerance` predicate, similar to `std::equal_to` et al, and use that with `BOOST_TEST(p(x,y))`. This would also make `BOOST_TEST_ALL_WITH` slightly easier to use.

We even have a sample implementation of [with_tolerance](https://github.com/breese/core/blob/develop/test/lightweight_test_all_with_test.cpp#L40) in the test suite.

HDembinski:
This is also nice. Perhaps all the implementations `test_eq_impl`, `test_lt_impl`, ... could be replaced with one generic backend. The predicate could then tweak the output. Or better we make a traits class that customizes the output based on the type of predicate.

With such a generic approach, one could also implement checks for relative and for absolute tolerance. Both are sometimes useful.

pdimov:
The `T` parameter is a bit of a problem because we probably don't want separate `BOOST_TEST_CLOSE` macros for each type. (And we can't use `decltype`.)

[...]

PR #39, opened 2018-08-06
new test backend based on predicates

HDembinski:
See #38 

The test backend uses predicates now, which reduces code duplication and makes the system more flexible. I implemented BOOST_TEST_CLOSE using this new flexibility (also added tests).

BOOST_TEST_CLOSE uses the predicate `close_to` suggested by @breese.


include/boost/core/close_to.hpp
```diff
@@ -0,0 +1,51 @@
...
+template <class T>
+struct close_to
+{
+    close_to(T relative_tolerance = std::numeric_limits<T>::epsilon(),
+             T absolute_tolerance = T(0))
+        : relative(relative_tolerance),
+          absolute(absolute_tolerance)
+    {}
+
+    bool operator() (const T& lhs, const T& rhs)
```
glenfe: `const`

include/boost/core/close_to.hpp
```diff
@@ -0,0 +1,51 @@
...
+namespace boost
+{
+
+namespace core
+{
+
+template <class T>
+struct close_to
+{
+    close_to(T relative_tolerance = std::numeric_limits<T>::epsilon(),
```
glenfe: `explicit`

include/boost/core/lightweight_test.hpp
```diff
@@ -118,106 +120,76 @@ template<class T> inline const void* test_output_impl(T volatile* v) { return co
...
-    {
-        BOOST_LIGHTWEIGHT_TEST_OSTREAM
-            << file << "(" << line << "): test '" << expr1 << " != " << expr2
-            << "' failed in function '" << function << "': "
-            << "'" << test_output_impl(t) << "' == '" << test_output_impl(u) << "'" << std::endl;
-        ++test_errors();
-    }
-}
+template <typename T>
+struct test_output_traits<std::equal_to<T> > {
```
glenfe: There are Boost libraries using Boost.Test even for facilities that need to support implementations where `BOOST_NO_TEMPLATE_PARTIAL_SPECIALIZATION` is defined.

include/boost/core/lightweight_test.hpp
```diff
@@ -118,106 +120,76 @@ template<class T> inline const void* test_output_impl(T volatile* v) { return co
...
-    if( t > u )
+    if( pred(t, u) )
     {
         report_errors_remind();
     }
     else
     {
         BOOST_LIGHTWEIGHT_TEST_OSTREAM
-            << file << "(" << line << "): test '" << expr1 << " > " << expr2
+            << file << "(" << line << "): test '" << expr1 << " " << test_output_traits<BinaryPredicate>::op() << " " << expr2
```
glenfe: Instead of `test_output_traits` the helper utility used here could also receive an output string.

pdimov:
It seems to me that converting the second argument to the type of the first would break `BOOST_TEST_GT('x', -1)`, `BOOST_TEST_EQ(nullptr, p)`, `BOOST_TEST_NE(&y, &x)` where `Y` derives from `X`, and so on.

include/boost/core/lightweight_test.hpp
```diff
@@ -118,106 +120,76 @@ template<class T> inline const void* test_output_impl(T volatile* v) { return co
...
-    }
-    else
-    {
-        BOOST_LIGHTWEIGHT_TEST_OSTREAM
-            << file << "(" << line << "): test '" << expr1 << " >= " << expr2
-            << "' failed in function '" << function << "': "
-            << "'" << test_output_impl(t) << "' < '" << test_output_impl(u) << "'" << std::endl;
-        ++test_errors();
-    }
+    test_with_impl(expr1, expr2, file, line, function, t, u, BinaryPredicate<T>());
```
glenfe: This is wrong when you pass in `std::equal_to` as the predicate. Before it would check `t == u` which could call `operator==(const T&, const U&)`. Now with your change it can't call such an operator, and would instead convert `U` to `T`.  You don't want `BinaryPredicate<T>` here, because you don't want `std::equal_to<T>`. You want `std:::equal_to<void>` but that isn't available until more recent C++ standards.

include/boost/core/lightweight_test.hpp
```diff
@@ -118,106 +120,76 @@ template<class T> inline const void* test_output_impl(T volatile* v) { return co
...
             << "' failed in function '" << function << "': "
-            << "'" << test_output_impl(t) << "' <= '" << test_output_impl(u) << "'" << std::endl;
+            << "'" << test_output_impl(t) << "' " << test_output_traits<BinaryPredicate>::anti_op() << " '" << test_output_impl(u) << "'" << std::endl;
         ++test_errors();
     }
 }
 
-template<class T, class U> inline void test_ge_impl( char const * expr1, char const * expr2,
-  char const * file, int line, char const * function, T const & t, U const & u )
+template<template <class> class BinaryPredicate, class T, class U>
```
glenfe: There's also no need to confine the predicates to only templates of the form `template<class>`

include/boost/core/lightweight_test.hpp
HDembinski: Ok, I could change these to functions overloads and thus avoid partial specialization.

include/boost/core/lightweight_test.hpp
HDembinski: I am not sure I understand.

glenfe:
@breese When or if a Boost library has a use for it, it can be moved out of the detail namespace, and then properly documented (and tested independently of the Lightweight test macros) as part of Boost.Core.

HDembinski:
I think the new version addresses all the concerns, let me know if I forgot something.

pdimov:
This looks sensible and is very close to what I envisaged as a design. Would you please rename the predicates along the lines of `lw_test_eq` instead of the generic `equal_to` so as not to pollute `detail`, and perhaps split this PR into two, one refactoring the existing functionality into predicate form, a separate one introducing `close_to`?

HDembinski:
Ok, will do. I was am aware of the "one-topic per PR" rule, but then considered the two patches to be so closely related, that I put them in one. Also to help with the discussion, so that it is clear where this is going.

HDembinski:
I suppose you also want a separate PR for line 209, where I corrected a mistake in the output print.

HDembinski:
Ok, that's a relief :)

pdimov:
As a matter of fact yes I do. -)

The reason to split separate things into their own PRs is not to adhere to a rule, but to allow them to be merged separately without being tied up together (and then reverted separately if need be.) This way straightforward changes can be applied immediately without waiting for the rest to crystallize.

Could you move the BinaryPredicate to be the first argument of `test_with_impl`? Thanks.

HDembinski:
@pdimov the predicate is also the last argument in `test_all_with_impl`. If I do as you suggest, it is inconsistent between `test_with_impl` and `test_all_with_impl`

HDembinski:
Otherwise I think I see why you request this. It is simply easier to see whether the correct predicate is used for a given BOOST_TEST... call when it is the first argument. But then we perhaps should also change `test_all_with_impl` in the same way.

pdimov:
Yes, I understand why you followed the style of the existing `test_all_with_impl`, but I still prefer the predicate first in this case. Let's leave `test_all_with_impl` as is.

pdimov:
Instead of `BinaryPredicate::op()`, `pred.op()` please. Makes no difference here, of course, but in principle, the `op` may depend on state. (The `anti_op` for `close_to(0.001)` could for instance return `!=[within 0.001]`.)

Could also rebase on develop to pick up the new tests I added.

HDembinski:
That is a great idea, I would like that for `close_to`.

HDembinski:
Thank you everyone for the merge and review!

PR #48, opened 2019-03-26
no_exceptions_support: Suppress conditional expression is constant warning

Kojoley:


pdimov:
I don't get this warning with msvc-14.0, 14.1, 14.2; only with 12.0 and below. So I wonder if the uglification is worth it.

Kojoley:
I agree that it looks ugly, but it does not cost you anything, and I do not think you look at the macro definition that often.

pdimov:
Is `_MSC_VER` (instead of `BOOST_MSVC`) intentional? It includes Borland (no idea if it's still alive), Clang, Intel.

Kojoley:
> Is `_MSC_VER` (instead of `BOOST_MSVC`) intentional? It includes Borland (no idea if it's still alive), Clang, Intel.

No special reason here. Probably did not spotted that `boost/config.hpp` is already included back than. You want me to replace it? The compilers that pretends with other usually implements or ignores those things.

> Borland (no idea if it's still alive)

AFAIK they ship old Clang with their own extensions now.

PR #50, opened 2019-04-18
print the qualified type

HDembinski:
`BOOST_TEST_TRAIT_SAME(T, T const)` produced the right result, but the wrong error message. It used to say
```
'int main()' ('T' != 'T')
```
This patch prints the correctly qualified type name
```
'int main()' ('T' != 'T const')
```

### DIFF
--- a/include/boost/core/lightweight_test_trait.hpp
+++ b/include/boost/core/lightweight_test_trait.hpp
@@ -20,6 +20,7 @@
 #include <boost/core/lightweight_test.hpp>
 #include <boost/core/typeinfo.hpp>
 #include <boost/core/is_same.hpp>
+#include <boost/config.hpp>
 
 namespace boost
 {
@@ -53,7 +54,7 @@ template<class T> inline bool test_trait_same_impl_( T )
 }
 
 template <class T>
-struct test_trait_stream_type {
+struct test_trait_print_type {
   static void stream() {
     BOOST_LIGHTWEIGHT_TEST_OSTREAM
       << boost::core::demangled_name( BOOST_CORE_TYPEID(T) );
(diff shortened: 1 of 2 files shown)


pdimov:
`stream` should be `print`; `T*` seems unnecessary, `T&&` is missing.

HDembinski:
A question: should this functionality go to core/typeinfo.hpp? It would be handy, for one thing, and even more, this line seems to compare the type names without qualifiers on some platforms, which would give wrong results: https://github.com/boostorg/core/blob/87026641885fdcea0d34439c10a9d80296664429/include/boost/core/typeinfo.hpp#L52

Putting the correct type name into core::typeinfo should fix this. I understand that core::typeinfo is supposed to be a drop-in replacement for std::typeinfo, so `core::typeinfo::name()` probably should still return the type name without qualifiers, but perhaps we could add an new method `core::typeinfo::full_name()` or something, which returns the correct type name.

HDembinski:
Looks like we are allowed to return a qualified name with the `name()` method, since there is basically no guarantee given regarding the name.
https://en.cppreference.com/w/cpp/types/type_info/name

HDembinski:
Thinking more about it, maybe this is the wrong approach to solve this issue. I am seeing this problem only because the `std::typeinfo::name()` method on my system doesn't show the qualifiers, but on another system it might. In this case, the current solution may return `T&&` for a `T&`, which would another bug.

A better and less platform-dependent solution would be the following:
```
template <class T> struct quote {};
#define BOOST_CORE_TYPEID(T) typeid(quote<T>)
```
Then when the name is returned, skip all characters up to the "quote<" string and remove the last ">". This can be done also for std::typeinfo, by inheriting from it.

pdimov:
`typeid(T)` strips references and cv qualifiers. See http://eel.is/c++draft/expr.typeid#4.

pdimov:
The `boost/config.hpp` include should come after the `core` includes.

HDembinski:
Why is it important to have the config header after the core headers?

pdimov:
It masks errors stemming from one of the headers using Config without including it.

HDembinski:
Ok, make sense.

glenfe:
I don't like that this uses `BOOST_LIGHTWEIGHT_TEST_OSTREAM` more than once for the same error expression.

In the expression, each  `boost::core::demangled_name(BOOST_CORE_TYPEID(X))` can just become `our_type<X>()` and the implementation becomes cleaner.

This also does not handle `const volatile`. 

See #54.

HDembinski:
I agree that @glenfe s version looks nicer.

pdimov:
Closing in favor of #54. Thanks Hans.

PR #52, opened 2019-04-19
fix warning of "unescapted special character"

HDembinski:
Running b2 in the doc folder gave the warning:
```
Jamfile.v2:52: Unescaped special character in argument <format>pdf:<xsl:param>boost.url.prefix=http://www.boost.org/doc/libs/release/libs/core/doc/html
```
This patch fixes the warning.


Lastique:
I think, it's better to envelop the URL in quotes.

Lastique:
Done in d9b524d.

PR #53, opened 2019-04-19
use type wrapper to get qualified type name

HDembinski:
This is an alternative for PR #50.


HDembinski:
Ok, nevermind, if typeid is supposed to return the unqualified type then this is wrong behavior.

PR #54, opened 2019-04-20
Print type with cvref qualifiers

glenfe:


HDembinski:
Is the `int` template parameter needed?

glenfe:
It's a workaround for earlier versions of VC.

HDembinski:
> It's a workaround for earlier versions of VC.

Wow. I am kinda glad that I joined only in the C++14 era.

PR #64, opened 2019-10-24
Add class version

HDembinski:
nvp was added to core, which is great to write serialize functions that do not depend directly on boost::serialization. but we also need boost::serialization::version in core, which is needed to inform boost serialization that the serialization format has changed.

The following patch has a simple implementation and some unit tests, but no documentation. There is some issue with the name. I followed the example of nvp and pulled serialization::version into the boost namespace, which yields boost::version. I would prefer if it was called boost::class_version, but I don't know how to make a templated alias without C++11 support.

### DIFF
--- a/include/boost/core/class_version.hpp
+++ b/include/boost/core/class_version.hpp
@@ -1,0 +1,42 @@
+//  Copyright (c) 2019 Hans Dembinski
+//
+//  Distributed under the Boost Software License, Version 1.0. (See
+//  accompanying file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+//  See http://www.boost.org/libs/core/doc/html/core/version.html for documentation.
+//
+#ifndef BOOST_CORE_CLASS_VERSION_HPP
+#define BOOST_CORE_CLASS_VERSION_HPP
+
+#include <boost/config.hpp>
+
+namespace boost {
+
+namespace core {
+template <int N>
+struct int_c {
+  typedef int value_type;
+  enum { value = N };
+  operator value_type() const { return N; }
+};
+} /* core */
+
+namespace serialization {
+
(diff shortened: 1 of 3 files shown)


pdimov:
Your patch doesn't make sense as written. It defines `boost::serialization::version` but tests `boost::version`.

What we could do is define `boost::class_version` and then derive `boost::serialization::version` from it for compatibility, although this could turn out a bit problematic if people start specializing both.

I'd advise against `int_c`. Just define `value` directly inside `version`, and drop `value_type` and the operator.

Lastique:
I don't like that we're adding bits of Boost.Serialization to Boost.Core. I'd rather we have Boost.SerializationCore for that, much like Boost.ThrowException.

HDembinski:
@pdimov You are right, it was late yesterday and I messed up the version that I pushed.

I used the approach you suggested to emulate an alias template, which is actually even better than a alias template now that I think about it. People who manually specialized boost::serialization::version will be ok, since boost.serialization will still use boost::serialization::version internally and it does not matter which version you specialize, assuming nobody will specialize both for the same type, which would be crazy.

I made `int_c` a separate class, because you sometimes want to not use the BOOST_CLASS_VERSION macro. When you want to version a templated class, you cannot use the macro. You have to manually write
```
template <class T>
struct my_class {};

// specializing class_version for all my_class
namespace boost {
template <class T>
struct class_version<my_class<T>> : ::boost::core::int_c<1> {};
}
```
This seems more leak less implementation details than
```
template <class T>
struct my_class {};

// specializing class_version for all my_class
namespace boost {
template <class T>
struct class_version<my_class<T>> {
   enum { value = 1 };
};
}
```

glenfe:
I explained to Hans why we don't want to do this (and shouldn't be making any pull requests without agreement from Robert, which we don't have for this anyway).

PR #66, opened 2019-11-23
Add deprecation warnings for headers in detail dir

Kojoley:


Kojoley:
I grepped boost for using those headers:

count | header
---- | -----------------------
1768 | detail/lightweight_test.hpp
29   | detail/iterator.hpp
23   | detail/no_exceptions_support.hpp
8    | detail/scoped_enum.hpp

and opt-outed of placing warning in `detail/lightweight_test.hpp` because of the obvious noise it will produce.

pdimov:
Any objections to applying this?

Lastique:
> Any objections to applying this?

Not from me. Assuming we're ok, since noone objected so far.

Issue #49, opened 2019-04-18
move implicit_cast to core?


HDembinski:
The boost.conversion library contains `implicit_cast`, which has a simple implementation, no external dependencies and seems useful to have in boost.core. It would also be nice to document the cast better, for me https://stackoverflow.com/questions/868306/what-is-the-difference-between-static-cast-and-implicit-cast was useful to read.

anshu-khare-design:
Hello sir,
I am Anshu Khare and I would like to implement down_cast in boost.core.
Could you please link me to relevant sources?

Lastique:
> Could you please link me to relevant sources?

What sources? This is a Boost.Core issue, this project contains Boost.Core sources.

Also, your post is not related to this issue. Please don't pollute the comments with off-topic.

anshu-khare-design:
Respected sir,
I am really sorry for my comment on the post. I will make sure that it is
not repeated again. This is the first time I am working on GitHub and open
source and I am a novice. I have done courses on C++ and I really wanted to
contribute to the organisation but when I cloned the core library in my
system, I found many cpp and hpp files in different folders. I couldn't get
which file I should look up to for writing the code. I even don't know the
data types which has been used in the library. I have built some programs
related to my courses I have done , locally by writing them from scratch
but i dont know how to build them after cloning projects from github. I
have eclipse installed in my system. I would be really grateful if u could
guide me regarding the same as I am not able to get the proper guidance. I
request you to help me out.
Thank you

Lastique:
These kind of questions should be asked on the mailing list.

anshu-khare-design:
I have worked on this.

glenfe:
@HDembinski I don't think we want to do this. Unlike say, `use_default`, this doesn't exist on public interfaces and is trivial enough to live in detail namespaces if a dependency on Conversion is not desirable. No dependency cycles are broken or would be avoided by relocating this facility.

PR #71, opened 2020-01-20
Added implicit cast to core library : issue #49

anshu-khare-design:
I have worked on issue #49  of core library.

### DIFF
--- a/implicit_cast.cpp
+++ b/implicit_cast.cpp
@@ -1,0 +1,33 @@
+// ConsoleApplication1.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#pragma once
+#ifndef HEADER_H_
+#define HEADER_H_
+
+#include <iostream>
+using namespace std;
+
+namespace boost {
+	template <typename T, typename R>
+	R impli_cast(T& arg1, R& arg2)
+	{
+		return (R)arg1;
+	}
+
+
+}
+
+
+#endif
+
+// Run program: Ctrl + F5 or Debug > Start Without Debugging menu
+// Debug program: F5 or Debug > Start Debugging menu
+
+// Tips for Getting Started: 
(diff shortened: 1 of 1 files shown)


mloskot:
WTF 😨

anshu-khare-design:
Implicit_cast typically means that we are able to convert one datatype to another.
If this is wrong, could you please link me to relevant sources?

mloskot:
> If this is wrong, could you please link me to relevant sources?

The relevant place to look is the `implicit_cast.cpp` file you added to the library in this PR. 
Look at it again and review what you have added, what it is, if/how it integrates with the library. 

To me, this is a useless file `.cpp` file, a randomly dropped in rubbish.

Answering those initial questions should already tell you it is a good idea to close this PR and forget, so I should not even make comments about quality of the code you submitted:

- useless comments
- not to mention `using namespace std`
- useless as unused `#include <iostream>`
- useless `#pragma once` in `.cpp` file, not to mention it is compiler-specific feature used without proper preprocessor conditional
- useless header guard used in `.cpp` file, not to mention the guard does not follow the https://www.boost.org/development/header.html

If you really really really die hard to contribute to `boostorg/core`, then at least try really really really hard to get familiar with the current contents of the library to see how it is structured, integrated, built (or perhaps it is not built, hint!), etc.

glenfe:
Everything about this is wrong. 

For your edification only, this is the implementation of `implicit_cast`:

```
template<class T>
BOOST_CONSTEXPR inline T implicit_cast(typename type_identity<T>::type v)
{
    return v;
}
```
We do not want to address #49, so I'm closing this PR.

mloskot:
@glenfe Shall the #49 be closed as well?

anshu-khare-design:
Actually that was my actual code..Actually typed it on on an online ide..Hence all that rubbish came in.

PR #73, opened 2020-04-08
Changes for Embarcadero C++ clang-based compilers, targeting Boost 1.74

eldiener:


.travis.yml
```diff
@@ -275,7 +275,7 @@ matrix:
 
     - os: linux
       compiler: clang++-10
-      env: TOOLSET=clang COMPILER=clang++-10 CXXSTD=03,11,14,17,20
+      env: TOOLSET=clang COMPILER=clang++-10 CXXSTD=03,11,14,17,2a
```
Lastique: Please, revert this. Boost.Build has been updated to support `cxxstd=20`.

eldiener: Yet the clang-10.0 test was failing because of the use of '20' instead of '2a'. That is why I made the change. Notice that all the other clang tests have '2a' and not '20' and succeeded. Are you saying that if I revert this the clang-10.0 test, which has '20' instead of '2a' will succeed when this PR is tested again ?

Lastique:
Merged just the essential commit.

.travis.yml
Lastique: clang 10 is the first version that supports `-std=c++20`, that's why the others use 2a. Boost.Build did not have support for `cxxstd=20` until https://github.com/boostorg/build/pull/592 was merged recently. If you restarted the CI, that error would not happen.

pdimov: > Boost.Build has been updated to support cxxstd=20.

Not on master, which means that when we merge to master we'll get Travis failures again.

Rene only merges to master when he does a release.

Lastique: Ok, I will change to 2a then.

PR #74, opened 2020-04-14
Implement allocator utilities

glenfe:
Instead of `allocator_traits` in Core, the same functionality provided by individual templates. This reflects recent C++ standard committee reflector discussion around how `allocator_traits` would be designed today instead.

### DIFF
--- a/doc/allocator_access.qbk
+++ b/doc/allocator_access.qbk
@@ -1,0 +1,244 @@
+[/
+Copyright 2020 Glen Joseph Fernandes
+(glenjofe@gmail.com)
+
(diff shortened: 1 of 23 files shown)


doc/allocator_access.qbk
```diff
@@ -0,0 +1,241 @@
...
+
+[section Examples]
+
+The following example shows these utilities used in the definition of
+an allocator-aware container class:
+
+```
+template<class T, class A = boost::default_allocator<T> >
+class container
+    : boost::emtpy_value<typename boost::allocator_rebind<A, T>::type> {
```
pdimov, line 38: emtpy

PR #77, opened 2020-06-06
Added BOOST_TEST_WITH macro to verify test results with custom predicate

breese:
Suggestion for #38.

Added BOOST_TEST_WITH macro. Moved operation name of built-in predicates to a traits class.


pdimov:
Thanks, I was thinking along the same lines. This will fail under C++03 but that's fixable and I can take care of it.

breese:
Please do

PR #82, opened 2020-11-10
Disable some of the calng-tidy warnings for the demangle.hpp header

apolukhin:


Lastique:
I really don't like cluttering the code with obscure comments to satisfy some tool.

glenfe:
Agree with @Lastique here.

apolukhin:
OK, closing :)

PR #92, opened 2021-07-01
Restore support for ancient Mac OS

evanmiller:
Mac OS 10.4 and earlier lack Availability.h, causing compilation to fail. Rearrange the macro logic to include Availability.h only if we already think there may be support for uncaught exceptions.


Lastique:
Thanks.

PR #95, opened 2021-08-11
Update GitHub Actions CI file

sdarwin:
The Ubuntu 16.04 environment is scheduled to be removed from GitHub Actions in September 2021. Migrate those jobs to Docker containers or Ubuntu 18.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,7 @@ jobs:
             install: g++-4.7
           - toolset: gcc-4.8
             cxxstd: "03,11"
-            os: ubuntu-20.04
-            container: ubuntu:16.04
+            os: ubuntu-18.04
             install: g++-4.8
           - toolset: gcc-4.9
             cxxstd: "03,11"
@@ -43,8 +42,7 @@ jobs:
             container: ubuntu:16.04
           - toolset: gcc-6
             cxxstd: "03,11,14,1z"
-            os: ubuntu-20.04
-            container: ubuntu:16.04
+            os: ubuntu-18.04
             install: g++-6
           - toolset: gcc-7
             cxxstd: "03,11,14,17"
@@ -90,20 +88,17 @@ jobs:
           - toolset: clang
             compiler: clang++-3.9
             cxxstd: "03,11,14"
-            os: ubuntu-20.04
-            container: ubuntu:16.04
+            os: ubuntu-18.04
             install: clang-3.9
           - toolset: clang
             compiler: clang++-4.0
             cxxstd: "03,11,14"
-            os: ubuntu-20.04
-            container: ubuntu:16.04
+            os: ubuntu-18.04
             install: clang-4.0
           - toolset: clang
             compiler: clang++-5.0
             cxxstd: "03,11,14,1z"
-            os: ubuntu-20.04
-            container: ubuntu:16.04
+            os: ubuntu-18.04
             install: clang-5.0
           - toolset: clang
             compiler: clang++-6.0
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -130,16 +142,31 @@ jobs:
...
 
     steps:
+      - name: Check if running in container
+        if: matrix.container != ''
+        run: echo "GHA_CONTAINER=${{ matrix.container }}" >> $GITHUB_ENV
+      - name: If running in container, upgrade packages
+        if: matrix.container != ''
+        run: |
+            apt-get -o Acquire::Retries=3 update && DEBIAN_FRONTEND=noninteractive apt-get -y install tzdata && apt-get -o Acquire::Retries=3 install -y sudo software-properties-common wget curl apt-transport-https make apt-file sudo unzip libssl-dev build-essential autotools-dev autoconf automake g++ libc++-helpers python ruby cpio gcc-multilib g++-multilib pkgconf python3 ccache libpython-dev
+            sudo apt-add-repository ppa:git-core/ppa
```
Lastique, line 156: Why install git from PPA?

.github/workflows/ci.yml
```diff
@@ -130,16 +142,31 @@ jobs:
...
+    container: ${{ matrix.container }}
 
     steps:
+      - name: Check if running in container
+        if: matrix.container != ''
+        run: echo "GHA_CONTAINER=${{ matrix.container }}" >> $GITHUB_ENV
+      - name: If running in container, upgrade packages
+        if: matrix.container != ''
+        run: |
+            apt-get -o Acquire::Retries=3 update && DEBIAN_FRONTEND=noninteractive apt-get -y install tzdata && apt-get -o Acquire::Retries=3 install -y sudo software-properties-common wget curl apt-transport-https make apt-file sudo unzip libssl-dev build-essential autotools-dev autoconf automake g++ libc++-helpers python ruby cpio gcc-multilib g++-multilib pkgconf python3 ccache libpython-dev
```
Lastique, line 155: Is `sudo` missing here?

.github/workflows/ci.yml
```diff
@@ -130,16 +142,31 @@ jobs:
...
+            sudo apt-add-repository ppa:git-core/ppa
+            sudo apt-get -o Acquire::Retries=3 update && apt-get -o Acquire::Retries=3 -y install git
+            python_version=$(python3 -c 'import sys; print("{0.major}.{0.minor}".format(sys.version_info))')
+            sudo wget https://bootstrap.pypa.io/pip/$python_version/get-pip.py
+            sudo python3 get-pip.py
+            sudo /usr/local/bin/pip install cmake
```
Lastique, line 161: Why install CMake using pip?

.github/workflows/ci.yml
```diff
@@ -188,7 +215,7 @@ jobs:
 
       - name: Run tests
         run: |
-          BUILD_JOBS=$((nproc || sysctl -n hw.ncpu) 2> /dev/null)
+          BUILD_JOBS=$( (nproc || sysctl -n hw.ncpu) 2> /dev/null )
```
Lastique, line 218: Is this necessary?

.github/workflows/ci.yml
sdarwin: Why install git from PPA? This CI script was designed to run on all the target operating systems including ubuntu LTS 12.04, 14.04, 16.04 etc. Those versions have quite out-of-date versions of git. This strategy, of installing newer utility packages such as git on older OS versions seems to be common with CI providers. Both Github Actions and Travis are doing that: they are providing the user with _newer_ versions of compilers, shells, etc.

.github/workflows/ci.yml
sdarwin: Is sudo missing here? Not on that line, no. The sudo package isn't pre-installed on all base docker images. It will be in the next step. After it's installed, then it can be used.

.github/workflows/ci.yml
sdarwin: Why install CMake using pip? Similar to the git PPA issue. This is a way to control the version of cmake, potentially choosing any version, and be able to install that latest copy on an older operating system.

.github/workflows/ci.yml
sdarwin: Is this small whitespace code change necessary? Yes. The script was crashing on Ubuntu 16 without it.  You could confirm, by reverting the change and then run CI.  I believe the reason it wasn't crashing in GitHub Actions before, is because they are providing _newer_ versions of certain packages, a modified environment.

Lastique: I have tried this in my own GHA attempts, and it still doesn't work. I believe, the problem is that `/bin/sh` is `dash` on Ununtu and older Ubuntu versions simply don't support the `$()` syntax, or at least not completely support.

I worked around this by forcing `bash` shell, as described [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun) (by setting `defaults.run.shell` to `bash`).

Lastique:
I have committed an updated config file that incorporates most of this PR and some other changes. Thank you for the submission.

Issue #98, opened 2021-10-31
Warnings: conditional expression is constant


vinniefalco:
>  boost/core/bit.hpp(310,1): warning C4127: conditional expression is constant

pdimov:
When?

vinniefalco:
When what?

PR #99, opened 2021-10-31
bit.hpp: Fix MSVC warnings about constant conditional expressions

Lastique:
Also use `if constexpr` where possible.

Closes https://github.com/boostorg/core/issues/98.

### DIFF
--- a/include/boost/core/bit.hpp
+++ b/include/boost/core/bit.hpp
@@ -31,6 +31,12 @@
 # endif
 #endif // defined(_MSC_VER)
 
+#if defined(BOOST_MSVC)
+# pragma warning(push, 3)
+// conditional expression is constant
+# pragma warning(disable: 4127)
+#endif // defined(BOOST_MSVC)
+
 namespace boost
 {
 namespace core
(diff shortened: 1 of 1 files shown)


glenfe:
Don't older MSVC versions (which do not have if-constexpr) also have this warning, though? I would have thought multiple overloads with enable-if would be the way to go here.

Lastique:
> Don't older MSVC versions (which do not have if-constexpr) also have this warning, though?

This is why I also disable the warning with a pragma.

> I would have thought multiple overloads with enable-if would be the way to go here.

`enable_if` is not necessary, a simple set of overloads with a tag argument would suffice. But that would be a more invasive change that I didn't feel like doing before 1.78. I can do that if that's the preferred way.

pdimov:
I suppose I can rewrite it with tag dispatch, but I don't see it as an improvement. I've disabled the warnings.

Andrey, since I enabled `warnings-as-errors`, I had to touch some of the tests you wrote. Feel free to revert those and apply a different suppression if you prefer.

Lastique:
Should I close this PR?

> Andrey, since I enabled `warnings-as-errors`, I had to touch some of the tests you wrote.

That's ok, though I generally don't support warnings as errors.

pdimov:
There's no sense in fixing warnings if we don't use warnings as errors in the tests. Every change inevitably has a chance of reintroducing some.

PR #101, opened 2021-11-01
bit.hpp: Use C++17 if constexpr where possible

Lastique:


Lastique:
@pdimov Ping?

pdimov:
I'm not sure why would I want this change. It makes the source uglier and does effectively nothing else. At the same time, I don't have any good arguments against, either.

Lastique:
It reduces the amount of code to compile. In each function, only one branch is enabled for a given call, the rest are just wasting compile time on overload resolution and DCE. `if constexpr` should allow to discard these branches earlier.

pdimov:
All right, let's give it a go.

PR #115, opened 2022-05-13
Remove inline friend begin and end functions

glenfe:


glenfe:
Going to let the tests cycle on this pull request. I'll add a test case tomorrow around the MSVC defect.

PR #122, opened 2022-08-19
add boost_core.natvis and interface project sources

vinniefalco:
moved the file to extra/ and tidied up the CML

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,13 @@ target_link_libraries(boost_core
     Boost::throw_exception
 )
 
+file(GLOB_RECURSE boost_core_IDEFILES CONFIGURE_DEPENDS include/*.hpp)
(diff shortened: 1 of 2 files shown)


glenfe:
Looks OK to me.

Lastique:
I'm ok too.

pdimov:
I'll merge this - eventually - but I want to think about the CML part a bit first.

vinniefalco:
Any thoughts on this?

pdimov:
I'm sitting on all these because while the .natvis files can be added immediately, I don't want to make the CMakeLists.txt changes in this manner because it will make them not match the `boostdep --cmake` output so they'll need to be maintained by hand from that point on, and I'd like to avoid that.

I was going to pick and merge the .natvis files by hand but I'm busy with other things now.

vinniefalco:
Can boostdep integrate .natvis files into the CML if it sees them?

PR #128, opened 2022-09-20
Add fclose_deleter

Lastique:
`fclose_deleter` can be used as a deleter function object for `std::FILE` pointers returned by `std::fopen`.

### DIFF
--- a/doc/core.qbk
+++ b/doc/core.qbk
@@ -60,6 +60,7 @@ criteria for inclusion is that the utility component be:
 [include noinit_adaptor.qbk]
 [include noncopyable.qbk]
 [include null_deleter.qbk]
+[include fclose_deleter.qbk]
 [include nvp.qbk]
 [include pointer_traits.qbk]
 [include quick_exit.qbk]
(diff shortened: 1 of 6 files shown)


pdimov:
I'm not quite sure this belongs here, even though we do have `null_deleter`. Which probably doesn't belong here either.

Lastique:
Should I revert?

pdimov:
Utility seems a better place for `fclose_deleter`, but having `null_deleter` here and `fclose_deleter` there wouldn't be particularly consistent. So maybe not.

You have to check `p` for `NULL` though. `unique_ptr` skips the deleter on `NULL`, but `shared_ptr` does not. And you should probably have a test where `fopen` returns `NULL`. Maybe test with `boost::shared_ptr` too.

Lastique:
> You have to check `p` for `NULL` though. `unique_ptr` skips the deleter on `NULL`, but `shared_ptr` does not.

Hmm, I thought `shared_ptr` also didn't call the deleter on `NULL`. Will fix, thanks.

PR #131, opened 2022-12-06
Fix for sprintf deprecation warning

mborland:
Use same method as https://github.com/boostorg/assert/commit/601ee4ba7ae9d9b532dab5f7e757f4187dce6300.

### DIFF
--- a/include/boost/core/lightweight_test.hpp
+++ b/include/boost/core/lightweight_test.hpp
@@ -202,6 +202,14 @@ inline unsigned long test_output_impl( char32_t const& v ) { return v; }
 #pragma warning(disable: 4996)
 #endif
 
+// Use snprintf if available as some complilers (clang 14.0) issue deprecation warnings for sprintf
+#if ( defined(_MSC_VER) && _MSC_VER < 1900 ) || ( defined(__MINGW32__) && !defined(__MINGW64_VERSION_MAJOR) ) || \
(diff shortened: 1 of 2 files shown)


mborland:
@pdimov Can you please approve the CI run since I am a first time contributor to this library? Thanks.

Lastique:
I'd prefer a function instead of a macro, something like [this](https://github.com/boostorg/log/blob/3f22dbe131908e6d72c47c9d17b221ddae873bfc/include/boost/log/detail/snprintf.hpp). I can make a PR moving that header from Boost.Log to Boost.Core.

And if it is a macro, at least make it private (i.e. `BOOST_CORE_DETAIL_SNPRINTF`).

mborland:
> Why is the condition here not the same as the one in Assert?

I saw that you have C++98 and 03 in the ci.yaml file, and `snprintf` is from C++11. If it is superfluous I can remove it.

pdimov:
I also have C++03 in Assert's CI.

The condition is technically correct, as snprintf is C99 and therefore C++11, but checking `__cplusplus` like this is true for every MSVC, because they still define it to 199711L.

Since the condition in Assert works, I assume that all supported compilers do provide `std::snprintf` even in C++98/C++03 mode.

Also, you should `#undef` the macro afterwards, as it's done in Assert.

mborland:
Pending further review this should be good to merge.

PR #132, opened 2022-12-06
Add portable snprintf/vsnprintf definition

Lastique:
This definitions is mostly a workaround for older MSVC versions that only provided non-portable `_snprintf` etc. that are not fully conforming to the standard `snprintf`. This implementation fixes its issues wrt. null termination and returned values in case of buffer overflows.

On platforms that support the standard `snprintf`, the definitions in the header are equivalent to the standard functions.

### DIFF
--- a/doc/changes.qbk
+++ b/doc/changes.qbk
@@ -1,10 +1,18 @@
 [/
   Copyright 2021 Peter Dimov
+  Copyright 2022 Andrey Semashev
   Distributed under the Boost Software License, Version 1.0.
   https://boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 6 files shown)


pdimov (approved):
Looks good at a glance.

pdimov:
Are you waiting for me to merge this? It would be useful f.ex. for https://github.com/boostorg/lexical_cast/pull/56.

Lastique:
I was going to merge it after the 1.81 release to avoid potential inconvenience in case if we need to merge develop to master for a pre-release bug fix. But I guess I can do it early.

PR #135, opened 2022-12-22
Deprecate `boost::core::is_same` and the associated header

Lastique:
Moved `is_same` implementation to detail (both directory and namespace) to use in the public headers and avoid introducing new dependencies. All other uses were ported to Boost.TypeTraits. Likewise, the documentation now recommends users to use Boost.TypeTraits or C++ standard library instead.

Also, removed unnecessary includes and added missing ones in a few places.

### DIFF
--- a/test/allocator_difference_type_test.cpp
+++ b/test/allocator_difference_type_test.cpp
@@ -6,8 +6,8 @@ Distributed under the Boost Software License, Version 1.0.
 (http://www.boost.org/LICENSE_1_0.txt)
 */
 #include <boost/core/allocator_access.hpp>
-#include <boost/core/is_same.hpp>
 #include <boost/core/lightweight_test_trait.hpp>
+#include <boost/core/detail/is_same.hpp>
 
 template<class T>
 struct A1 {
@@ -22,9 +22,9 @@ struct A2 {
 
 int main()
 {
-    BOOST_TEST_TRAIT_TRUE((boost::core::is_same<short,
+    BOOST_TEST_TRAIT_TRUE((boost::core::detail::is_same<short,
         boost::allocator_difference_type<A1<char> >::type>));
-    BOOST_TEST_TRAIT_TRUE((boost::core::is_same<std::ptrdiff_t,
+    BOOST_TEST_TRAIT_TRUE((boost::core::detail::is_same<std::ptrdiff_t,
         boost::allocator_difference_type<A2<int> >::type>));
     return boost::report_errors();
 }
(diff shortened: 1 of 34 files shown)


test/allocator_difference_type_test.cpp
```diff
@@ -6,8 +6,8 @@ Distributed under the Boost Software License, Version 1.0.
 (http://www.boost.org/LICENSE_1_0.txt)
 */
 #include <boost/core/allocator_access.hpp>
-#include <boost/core/is_same.hpp>
 #include <boost/core/lightweight_test_trait.hpp>
+#include <boost/type_traits/is_same.hpp>
```
glenfe, line 10: I would prefer if these unit tests in Core used boost::core::detail::is_same instead of requiring TypeTraits

pdimov: Or maybe just change to BOOST_TEST_TRAIT_SAME.

Lastique: Boost.TypeTraits is already required by some tests, so you're not gaining anything by using `core::detail::is_same`. I chose Boost.TypeTraits because that is what we recommend in docs. But I can change it if you want.

Lastique: I thought about `BOOST_TEST_TRAIT_SAME` but it seems to require variadic macros, i.e. C++11.

pdimov: Variadic macros are supported as an extension by all compilers we care about, see https://github.com/boostorg/core/blob/develop/test/lightweight_test_test4.cpp.

Lastique: I know at least clang issues warnings about using variadic macros in C++03 mode. I think, recent gcc versions do too. I'd rather not mess with warnings.

Lastique: @glenfe Done. The `allocator_size_type_test.cpp` test I left using `boost::is_same` because it needs `boost::make_unsigned` anyway.

pdimov: See https://github.com/boostorg/core/blob/75c765cc1333e7656c52feddace1f303e53419a3/test/Jamfile.v2#L121-L122, https://github.com/boostorg/core/blob/75c765cc1333e7656c52feddace1f303e53419a3/test/Jamfile.v2#L100-L103, and https://github.com/boostorg/core/blob/75c765cc1333e7656c52feddace1f303e53419a3/include/boost/core/lightweight_test_trait.hpp#L84-L87.

glenfe (approved):
Looks good to me

pdimov (approved):

PR #137, opened 2023-01-06
Fix error: 'ref_empty' is not a member

denzor200:


pdimov:
This doesn't look like the right fix. But in either case, please first give an example that fails.

denzor200:
> Look at the CI fails in develop

https://github.com/boostorg/pfr/actions/runs/3857066454/jobs/6574035091
It also breaks Boost PFR's CI and stops my work on it.

pdimov:
Oh, @Lastique has been doing things. Andrey, please do things on a branch first, before merging to develop.

Lastique:
Oops, sorry for this. I thought I ran the tests locally, but apparently I did so on an intermediate version of the code. Should be fixed now.

Lastique:
> Andrey, please do things on a branch first, before merging to develop.

Ok. Sorry again.

PR #144, opened 2023-04-25
Implement latch

cmazakas:
Add an implementation of C++20's `std::latch`

https://en.cppreference.com/w/cpp/thread/latch

Note, notifying on the condition variable without the lock being held triggers a "dubious" warning from helgrind.

### DIFF
--- a/doc/core.qbk
+++ b/doc/core.qbk
@@ -83,3 +83,4 @@ criteria for inclusion is that the utility component be:
 [include uncaught_exceptions.qbk]
 [include use_default.qbk]
 [include verbose_terminate_handler.qbk]
+[include latch.qbk]
--- a/doc/latch.qbk
+++ b/doc/latch.qbk
@@ -1,0 +1,164 @@
+[/
+  Copyright 2023 Christian Mazakas
+  Distributed under the Boost Software License, Version 1.0.
+  https://boost.org/LICENSE_1_0.txt
+]
+
+[section:latch latch]
+
+[simplesect Authors]
+
+* Christian Mazakas
+
+[endsimplesect]
+
+[section Header <boost/core/latch.hpp>]
+
+The header `<boost/core/latch.hpp>` implements, in a portable way,
+the C++20 `<latch>` header.
+
+`latch` is a single-use barrier which counts downwards, enabling synchronization
+between a set of threads. The counter can be manually decremented by any value,
+but decrementing below zero produces UB. The maximum number of waiters is
+specified to be `boost::core::latch::max()`.
+
+[endsect]
+
+[section Example]
+
+```
+std::ptrdiff_t const num_threads = 16;
+boost::core::latch l(num_threads);
+
+std::vector<std::thread> threads;
+for (int i = 0; i < num_threads; ++i) {
+  threads.emplace_back([&l] {
+    // block until all threads have reached this statement
+    l.arrive_and_wait();
+  });
+}
+
+for (auto& t: threads) { t.join(); }
+```
+
+[endsect]
+
+[section Reference]
+
+```
+namespace boost
+{
+namespace core
+{
+
+class latch
+{
+    explicit latch(std::ptrdiff_t expected);
+
+    latch(latch const &) = delete;
+    latch &operator=(latch const &) = delete;
+
+    ~latch() = default;
+
+    void count_down(std::ptrdiff_t n = 1);
+
+    bool try_wait() const noexcept;
+
+    void wait() const;
+
+    void arrive_and_wait(std::ptrdiff_t n = 1);
+
+    static constexpr std::ptrdiff_t max() noexcept;
+};
+
+} // namespace core
+} // namespace boost
+```
+
+[section Constructors]
+
+[variablelist
+  [
+    [`explicit latch(std::ptrdiff_t expected);`]
+    [
+      [variablelist
+        [[Constraints][`expected >= 0 && expected <= boost::core::latch::max()`.]]
+        [
+          [Effects]
+          [Constructs a latch with an internal counter value of `expected`.]
+        ]
+      ]
+    ]
+  ]
+  [
+    [`latch(latch const &) = delete;`]
+    [
+      [variablelist
+        [[Constraints][`latch` is not copyable or movable.]]
+      ]
+    ]
+  ]
+]
+
+[endsect]
+
+[section Member Functions]
+
+[variablelist
+  [
+    [`void count_down(std::ptrdiff_t n = 1);`]
+    [
+      [variablelist
+        [[Constraints][`n` must not be larger than the current internal count.]]
+        [[Effects][Decrements the internal counter by `n`.]]
+      ]
+    ]
+  ]
+  [
+    [`bool try_wait() const noexcept;`]
+    [
+      [variablelist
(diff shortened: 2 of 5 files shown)


doc/core.qbk
```diff
@@ -83,3 +83,4 @@ criteria for inclusion is that the utility component be:
 [include uncaught_exceptions.qbk]
 [include use_default.qbk]
 [include verbose_terminate_handler.qbk]
+[include latch.qbk]
```
pdimov, line 86: These are in alphabetical order.

doc/latch.qbk
```diff
@@ -0,0 +1,164 @@
...
+
+[endsect]
+
+[section Example]
+
+```
+std::ptrdiff_t const num_threads = 16;
+boost::core::latch l(num_threads);
+
+std::vector<std::thread> threads;
```
pdimov, line 33: It will be better if the example compiles instead of being a fragment. So it would be better if this all is in a function.

doc/latch.qbk
```diff
@@ -0,0 +1,164 @@
...
+
+* Christian Mazakas
+
+[endsimplesect]
+
+[section Header <boost/core/latch.hpp>]
+
+The header `<boost/core/latch.hpp>` implements, in a portable way,
+the C++20 `<latch>` header.
+
```
pdimov, line 19: The C++11 requirement should be mentioned explicitly.

Lastique:
IMO, this needs to go to Boost.Thread, not Boost.Core. Actually, [there already is one](https://github.com/boostorg/thread/blob/52b62ee7b029028126c4eaf36e377833b0666510/include/boost/thread/latch.hpp), so I'm inclined to close this.

include/boost/core/latch.hpp
```diff
@@ -0,0 +1,104 @@
...
+//  Distributed under the Boost Software License, Version 1.0.
+//  https://www.boost.org/LICENSE_1_0.txt
+
+#ifndef BOOST_CORE_LATCH_HPP_INCLUDED
+#define BOOST_CORE_LATCH_HPP_INCLUDED
+
+#include <boost/config.hpp>
+
+#if defined(BOOST_NO_CXX11_HDR_CONDITION_VARIABLE)
+BOOST_PRAGMA_MESSAGE("boost::core::latch requires <condition_variable>")
```
pdimov, line 11: These messages are unnecessary. The inclusion of `<condition_variable>` below is enough to imply the this header requires it.

doc/latch.qbk
```diff
@@ -0,0 +1,164 @@
...
+        [[Constraints][`n` must not be larger than the current internal count.]]
+        [[Effects][Decrements the internal counter by `n` and if the counter non-zero, blocks the current thread.]]
+      ]
+    ]
+  ]
+  [
+    [`static constexpr std::ptrdiff_t max() noexcept;`]
+    [
+      [variablelist
+        [[Effects][Returns an implementation-defined number representing the maximum amount of waiters. Currently INT_MAX.]]
```
pdimov, line 154: Why not PTRDIFF_MAX?

pdimov:
Because otherwise we'll duplicate it in the test suites of all the libraries that need it.

Lastique:
No, what I'm asking is why the tests can't use the implementation in Boost.Thread? If there's a problem with it, let's fix it there.

pdimov:
A dependency on Boost.Thread just for this simple component is hardly justified.

Lastique:
IMHO, there's nothing wrong with the dependency. The duplication is far less justified.

Why do we write libraries in the first place, when we ourselves don't want to use them??

cmazakas:
> IMHO, there's nothing wrong with the dependency. The duplication is far less justified.
> 
> Why do we write libraries in the first place, when we ourselves don't want to use them??

Well, one thing to keep in mind is that Core is slowly but surely growing into a C++17/20 polyfill library.

Core already contains `std::span`, `<bit>` and has a `string_view` in its `detail` folder.

Thinking about this as a vcpkg user, it's quite nice I'd be able to just install boost-core instead of boost-thread which is a heavier weight dependency to polyfill a C++20 stdlib feature.

cmazakas:
> Re. `string_view`, I always was opposed to its addition.

This is interesting to me.

C++11 and 14 are still pretty popular standards in use today. `string_view` is a pretty great type that can exist easily in these versions of C++.

Right now, I'm not aware of any other maintained/reliable implementations of `string_view` in Boost. In this same vein, it seems like Thread hasn't received much attention in recent years.

Whatever intentions began for Core, the general evolution is tending towards implementing the small goodies from later C++ standards. `latch` fits this bill. It's simple, straight-forward and solves subtle implementation issues that people who hand-roll are likely to create.

Lastique:
> Right now, I'm not aware of any other maintained/reliable implementations of `string_view` in Boost.

The canonical implementation is in Boost.Utility.

Anyway, if other maintainers decide that this should be accepted, so be it. @glenfe, could you comment?

glenfe:
If this is only going to be used in the tests, I'd be inclined to just just keep using the one in Boost.Unordered (if not the one in Boost.Thread). 

If the one in Boost.Thread is going to be replaced with this (e.g. Thread dropping C++03, and aliasing boost::core::latch) then this being in Core is better.

pdimov:
The latter is probably not happening. Thread doesn't have a maintainer and is way too complex for drive-by maintenance.

pdimov:
With subsequent C++ standards adding nice things, we do have a need for a place where we put our "polyfills" when they meet certain criteria: relatively simple, no dependencies beyond the necessary (Assert and ThrowException), no deviation from standard semantics so no need for design reviews.

There's no better place than Core at the moment, but it might indeed not be the right one.

Maybe we need to have a separate repo/library, Boost.Compat, where these can go. (We already have Compatibility but it's obsolete and the name is a bit too long for my taste.)

glenfe:
Boost.Compat sounds like a good idea. 

Moreover, if we decide it is never for public consumption (e.g. types in there only used in implementation details, never on interfaces, or used in unit tests, ideally with a namespace that conveys the same) then we don't have to even think about a formal review for such a repository.

pdimov:
Sounds appealing but I think it would definitely be consumed by the public, so we should acknowledge it formally (and have proper docs).

The use in interface is an interesting point though. I suppose you're thinking of e.g. `span`, which would therefore not belong there.

glenfe:
Yes (which is shared by Histogram and Beast right now and used on the interfaces of both).

cmazakas:
Should we temporarily put a pin in this PR and instead directly address the mailing list about Boost.Compat?

I will incorporate Peter's feedback first, of course.

Kojoley:
> No, what I'm asking is why the tests can't use the implementation in Boost.Thread? If there's a problem with it, let's fix it there.
> IMHO, there's nothing wrong with the dependency. The duplication is far less justified.

1. Being in Boost.Thread makes it look like it requires using threads from Boost.Thread.
2. There is a problem with the dependency on Boost.Thread -- it requires compilation and also drags even more dependencies. In CMake it could be solved by providing separate 'library'.

I actually was under impression that the Core library is exactly for such things. Citing the readme:

> Boost.Core, part of collection of the [Boost C++ Libraries](https://github.com/boostorg), is a collection of core utilities used by other Boost libraries. The criteria for inclusion is that the utility component be:
> * simple,
> * used by other Boost libraries, and
> * not dependent on any other Boost modules except Core itself, Config, Assert, Static Assert, or Predef.

Issue #148, opened 2023-07-05
core/swap.hpp(81,1): fatal error C1202: recursive type or function dependency context too complex


afabri:
We get a ` boost_1_82_0\boost/core/swap.hpp(81,1): fatal error C1202: recursive type or function dependency context too complex`  when compiling [this program](https://gist.github.com/afabri/1d772a2807459ad07e85450f6c7477af)  with VC2019 or VC2022 and stdc++17 enabled.     [link](https://github.com/boostorg/core/blob/develop/include/boost/core/swap.hpp#L81)
Neither g++ nor clang have a problem. 
The code is nonsensical and we only boiled down the real code to make it a minimal example. Also ` #include <boost/swap.hpp>`  is not used but only needed for getting the error. 
Maybe this is more a VC++ issue than a boost issue.

afabri:
Hi @Lastique 
Do you by any chance have access to the VC++ compiler?  I ask because your commit concerning the noexception triggers the problem.

Lastique:
> Do you by any chance have access to the VC++ compiler?

Not at the moment or in the coming few weeks.

If you are willing to try something, you could try the following:

1. Envelop `boost::swap` definition [here](https://github.com/boostorg/core/blob/216999e552e7f73e63c7bcc88b8ce9c179bbdbe2/include/boost/core/swap.hpp#L77-L84) in a new namespace within namespace `boost`, say `namespace swap_adl_block`.
2. After that namespace closes, import `boost::swap_adl_block::swap` into namespace `boost` by adding `using swap_adl_block::swap` declaration.
3. Try if it fixes your problem.

pdimov:
Reduced example:

```
#include <boost/swap.hpp> // compiles if removed
#include <boost/checked_delete.hpp>

template<class T> class X
{
    X( X const& );

public:

    X() {}
};

int main()
{
    X< boost::checked_deleter<int> > x;
    return noexcept( swap( x, x ) );
}
```

Fails under all compilers (e.g. https://godbolt.org/z/ea65PWz9b). The reason is that `std::swap` is SFINAEd away when the type isn't move constructible, so the unqualified `swap` call ends up calling `boost::swap` recursively, forever.

pdimov:
Why did we need the change in https://github.com/boostorg/core/commit/8a8738a981e1d46146939f93ac0d751b827e7704?

Lastique:
On July 9, 2023 8:28:34 PM Peter Dimov ***@***.***> wrote:

[...]

PR #149, opened 2023-07-11
Rename `boost::swap` to `boost::core::invoke_swap`, deprecate `boost::swap`

Lastique:
The rename allows to avoid forming an infinite recursion in compile time (because of `noexcept` specification that needs to resolve the unqualified call to `swap`) or run time (in case if the `boost::swap` function is the only one suitable for the passed arguments).

To avoid the compile-time recursion, removed `noexcept` specification from `boost::swap`. The specification is still present in `boost::core::invoke_swap`.

Deprecated `boost::swap` and associated headers. `boost::core::invoke_swap` is defined in a new `boost/core/invoke_swap.hpp` header.

Updated docs and tests. Removed tests that check inclusion of deprecated headers.

Fixes https://github.com/boostorg/core/issues/148.


pdimov:
This looks good to me. @glenfe?

Lastique:
Also, do we want this to be part of 1.83? I did not create any PRs updating other libraries to switch over to `invoke_swap`.

pdimov:
It's probably too late for 1.83.

Lastique:
> It's probably too late for 1.83.

Yeah, I thought so too. I'll merge this after the release then. 1.83 will be released as currently is.

glenfe:
Looks good to me.

We might want to delay 1.83 for other reasons (Context submodule being disconnected from master HEAD has resulted in other issues) so I want to think about getting this in. Versus shipping 1.83 with broken Context etc.

PR #159, opened 2023-11-25
[bit] Add support for `__builtin_bswap16`

Lastique:


pdimov:
There's no need for that; all compilers recognize the pattern and turn it into a bswap16 if possible.

include/boost/core/bit.hpp
```diff
@@ -47,12 +47,19 @@
 # if __has_builtin(__builtin_bit_cast)
 #  define BOOST_CORE_HAS_BUILTIN_BIT_CAST
 # endif
+# if __has_builtin(__builtin_bswap16)
+#  define BOOST_CORE_HAS_BUILTIN_BSWAP16
+# endif
 #endif
 
-#if defined(BOOST_MSVC) && BOOST_MSVC >= 1926
```
pdimov, line 52: Why is this change needed?

include/boost/core/bit.hpp
```diff
@@ -47,12 +47,19 @@
...
+#if !defined(BOOST_CORE_HAS_BUILTIN_BIT_CAST) && (defined(BOOST_MSVC) && BOOST_MSVC >= 1926)
```
pdimov, line 55: Why is this change needed?

Lastique: The `BOOST_CORE_HAS_BUILTIN_BIT_CAST` macro could be defined above, if MSVC supports `__has_builtin` one day. It's just a sanity check to avoid redefinition of the macro and a potential warning.

It's not essential to the bswap16 change and hence is done in a separate commit. I can remove it if you want.

Lastique:
> There's no need for that; all compilers recognize the pattern and turn it into a bswap16 if possible.

I'd rather be on the safe side and explicitly say what we want. The intrinsic is there, so why not use it.

pdimov:
It's not available on all targets and can be non-constexpr, so it's not an automatic "win".

Lastique:
> It's not available on all targets and can be non-constexpr, so it's not an automatic "win".

Do you have an example of that? My understanding is that it is equivalent to bswap32/64, only in case of gcc it appeared a little later than the other two. Clang supported all three all along.

pdimov:
Looks like it wasn't supported on GCC 4.7 and Clang 3.1, but this shouldn't be an issue because GCC 4.7 doesn't have `__has_builtin` so it's not going to get used anyway. (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52624)

Appears to be constexpr when supported too.

Lastique:
Right, that's why I'm enabling it only since gcc 4.8.

include/boost/core/bit.hpp
```diff
@@ -47,12 +47,19 @@
...
 #  define BOOST_CORE_HAS_BUILTIN_BIT_CAST
 #endif
 
+#if !defined(BOOST_CORE_HAS_BUILTIN_BSWAP16) && (defined(BOOST_GCC_VERSION) && BOOST_GCC_VERSION >= 40800)
```
pdimov, line 59: Why BOOST_GCC_VERSION instead of BOOST_GCC?

pdimov: Well? This looks OK to merge except for this bit.

As far as I can see, `BOOST_GCC_VERSION` is mostly equivalent to `BOOST_GCC`, but is undocumented.

Seems to me that if we want "real GCC 4.8 or above", this should be `BOOST_GCC`, and if we want "GCC-4.8-compatible", this should be `__GNUC__ * 100 + __GNUC_MINOR__ >= 408`.

PR #165, opened 2024-02-07
Move checked deleters into internal namespace to block unintended ADL

Lastique:
This prevents adding namespace boost to ADL when the deleters are used in template parameters, e.g. in `std::unique_ptr`.


Lastique:
@pdimov Any objections?

pdimov:
It will break forward declarations of `checked_deleter`, but those should be rare, if they exist at all. So no.

PR #169, opened 2024-03-16
Fix deprecated warning ignores in allocator_access.hpp

### LINKED ISSUE #168: `boost::allocator_destroy()` emitting deprecated warnings for `std::pmr::polymorphic_allocator` in Clang 18.1.0 with C++20
```
#include <boost/core/allocator_traits.hpp>
#include <memory_resource>

int main()
{
    std::pmr::polymorphic_allocator<int> alloc;

    int* p = boost::allocator_allocate(alloc, 1);
    boost::allocator_destroy(alloc, p);
}
```
Compiled in `x86-64 clang 18.1.0` with `-std=c++20 -Werror`

https://godbolt.org/z/bzTGYqjEj

```
/opt/compiler-explorer/libs/boost_1_84_0/boost/core/allocator_access.hpp:567:7: error: 'destroy<int>' is deprecated: use 'allocator_traits::destroy' instead [-Werror,-Wdeprecated-declarations]
  567 |     a.destroy(p);
      |       ^
```

k3DW:
Closes #168


include/boost/core/allocator_access.hpp
```diff
@@ -43,9 +43,17 @@ _LIBCPP_SUPPRESS_DEPRECATED_PUSH
...
-#pragma warning(disable:4996)
+#if defined(__clang__) && defined(__has_warning)
+# if __has_warning("-Wdeprecated-declarations")
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wdeprecated-declarations"
+# endif
+#elif defined(_MSC_VER)
+# pragma warning(push)
+# pragma warning(disable: 4996)
+#elif defined(__GNUC__) && !(defined(__INTEL_COMPILER) || defined(__ICL) || defined(__ICC) || defined(__ECC)) && (__GNUC__ * 100 + __GNUC_MINOR__) >= 406
```
Lastique, line 54: We have [`BOOST_GCC`](https://www.boost.org/doc/libs/1_84_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.boost_informational_macros) macro for this.

pdimov:
I don't think the GCC part is needed; we don't get any warnings from GCC.

k3DW:
I'll remove the GCC portion

k3DW:
Some of the drone GCC stages are showing failures with `deprecated-declarations`, so I'll add back the GCC portion, this time using `BOOST_GCC`

test/Jamfile.v2
```diff
@@ -265,6 +265,7 @@ run allocator_destroy_test.cpp ;
 run allocator_construct_n_test.cpp ;
 run allocator_destroy_n_test.cpp ;
 run allocator_traits_test.cpp ;
+run allocator_pmr_test.cpp ;
```
Lastique, line 268: The test doesn't seem to execute any checks. Perhaps, it should be `compile` instead of `run`?

pdimov: Or maybe `link`.

Lastique: And while at it, if it doesn't need to execute any checks then it doesn't need to depend on `lightweight_test.hpp`.

k3DW: With `link` it looks like I still need a `main()` function, but I can leave it out with `compile`. I'll change it to `compile` to start

PR #186, opened 2025-01-04
Add reproducer for Clang < 15 bug with string_view & C++11 char types

Flamefire:


pdimov:
I see no CI failures for Clang 13/14 here.

Flamefire:
Seems using libstdc++ 12 and up requires Clang 15. Using libstdc++ 11 works also locally.

I'd say the test still makes sense to find potential regressions with the unicode char types.   
And the workaround in #185 doesn't make anything worse:  `end() == data() + size()` which the `std::basic_string` ctor does too if `data(), size()` is passed. But it helps in one (even if now less relevant) case:

PR #189, opened 2025-02-24
Use BOOST_ASSERT instead of a private assert macro for span

pdimov:
Instead of removing assertions under GCC 4.x to retain constexpr support, this retains the assertions and removes constexpr support under GCC 4.x instead. The presumption is that GCC 4.x users care more about assertions than they care about constexpr (which is hardly useful there anyway.)


glenfe:
I would rather just drop GCC 4 support in span instead.

pdimov:
One does not preclude the other.

glenfe:
Adding the asserts in the way I did was a non-breaking change. The first breaking change I want to make is dropping GCC 4 support. I'm happy to make it sooner.

pdimov:
It's not a breaking change if it doesn't break anything. :-)

pdimov:
Regardless of whether this is applied, develop needs to be merged to master.

> (Until then I'm fine with GCC 4 users not having asserts)

That's fine in principle, but in this specific case, `span` has only one purpose in life and it's to assert. :-)

PR #191, opened 2025-04-03
Initial support for the C++20 modules

apolukhin:
The PR follows an experimental modules approach from Boost.PFR https://github.com/boostorg/pfr/commit/60391652fa92b7a237583f00c863caa78ab8a0f1

The approach is an experiment. Any comments, suggestions are welcome!

### DIFF
--- a/include/boost/core/detail/is_same.hpp
+++ b/include/boost/core/detail/is_same.hpp
@@ -11,6 +11,8 @@
 
 #include <boost/config.hpp>
 
+#include <boost/core/detail/module_macro.hpp>
+
 #if defined(BOOST_HAS_PRAGMA_ONCE)
 # pragma once
 #endif
@@ -22,6 +24,8 @@ namespace core
 namespace detail
 {
 
+BOOST_CORE_BEGIN_MODULE_EXPORT
+
 template< class T1, class T2 > struct is_same
 {
     BOOST_STATIC_CONSTANT( bool, value = false );
@@ -31,6 +35,8 @@ template< class T > struct is_same< T, T >
 {
     BOOST_STATIC_CONSTANT( bool, value = true );
 };
+
+BOOST_CORE_END_MODULE_EXPORT
 
 } // namespace detail
 } // namespace core
--- a/module/core.cppm
+++ b/module/core.cppm
@@ -1,0 +1,207 @@
+// Copyright (c) 2025 Antony Polukhin
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// To compile manually use a command like the folowing:
+// clang++ -I ../include -std=c++20 --precompile -x c++-module core.cppm
+
+#define BOOST_CORE_BEGIN_MODULE_EXPORT export {
+#define BOOST_CORE_END_MODULE_EXPORT }
+
+module;
+
+#include <boost/config/header_deprecated.hpp>
+#include <boost/config.hpp>
+#include <boost/config/no_tr1/memory.hpp>
+#include <boost/config/pragma_message.hpp>
+#include <boost/config/workaround.hpp>
+
+#include <boost/assert.hpp>
+#include <boost/assert/source_location.hpp>
+
+#include <boost/current_function.hpp>
+#include <boost/static_assert.hpp>
+
+#include <boost/throw_exception.hpp>  // TODO: modularize
+
+#if (defined(__cpp_lib_uncaught_exceptions) && __cpp_lib_uncaught_exceptions >= 201411)
+#if defined(__APPLE__)
+#include <Availability.h>
+#endif
+#endif
+
+#ifndef BOOST_CORE_HAS_STD_MODULE
+#include <algorithm>
+#include <array>
+#include <cctype>
+#include <climits>
+#include <cmath>
+#include <cstdarg>
+#include <cstddef>
+#include <cstdio>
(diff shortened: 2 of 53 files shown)


include/boost/core/detail/is_same.hpp
```diff
@@ -22,6 +24,8 @@ namespace core
 namespace detail
 {
 
+BOOST_CORE_BEGIN_MODULE_EXPORT
```
Lastique, line 27: I don't think this should be exported. `core::detail::is_same` is an implementation detail and should not be used by users.

module/core.cppm
```diff
@@ -0,0 +1,207 @@
...
+#include <new>
+#include <stdexcept>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string>
+#include <time.h>
+#include <typeinfo>
+#include <type_traits>
+#include <utility>
+#include <wchar.h>
```
Lastique, line 64: Why are all these includes needed? Everything needed by Boost.Core should be included by Boost.Core headers.

module/core.cppm
```diff
@@ -0,0 +1,207 @@
...
+#include <boost/swap.hpp>
+#include <boost/iterator.hpp>
+#include <boost/utility/enable_if.hpp>
+#include <boost/utility/explicit_operator_bool.hpp>
+#include <boost/utility/swap.hpp>
+#include <boost/utility/addressof.hpp>
+#include <boost/noncopyable.hpp>
+}
+#else
+#include <boost/visit_each.hpp>
```
Lastique, line 144: Why the duplicate list of includes?

include/boost/core/detail/is_same.hpp
apolukhin: I also do not think that this should be exported.

However clang complains on exporting `namespce core {  using boost::core::detail::is_same; } `, so there's actually no choice if boost::core::is_same  is expprted

module/core.cppm
apolukhin: That is true for headers. But when a module is created, we need to place all the standard library and OS stuff into a global module fragment to avoid same symbols being ambiguously exported from different modules of Boost.

To do so I put all the headers right after the `module;`, so they end up in global module fragment. Later, after the `export module Boost.Core; ` I include all the library heders. Include guards of the standard library and OS stuff do they work, so the library headers do not actually include them again. As a result, only the library classes, functions and aliases apear after `export module Boost.Core; `, so the module exports only the library stuff

module/core.cppm
```diff
@@ -0,0 +1,207 @@
...
+#include <time.h>
+#include <typeinfo>
+#include <type_traits>
+#include <utility>
+#include <wchar.h>
+#endif
+
+export module Boost.Core;
+
+#ifdef BOOST_CORE_HAS_STD_MODULE
```
apolukhin, line 69: This macro should be changed to some Boost.Config macro that is defined if the compiler understands modules and there is module std. The includes in the headers should be guarded by that macro and use `import std; ` if it is avalable

include/boost/core/detail/is_same.hpp
Lastique: That `using`-declaration is for backward compatibility and deprecated, it should not be exported either but rather removed.

module/core.cppm
apolukhin: Now I also have a feeling that I'm doing something wrong. I'll try another approach soon

pdimov:
@anarthal 

https://anarthal.github.io/cppblog/modules
https://anarthal.github.io/cppblog/modules2
https://anarthal.github.io/cppblog/modules3

anarthal:
?

anarthal:
Oh, I misread this. @apolukhin you might have a look at the articles pointed out by Peter and the discussion we had on the ML. I will have a look at the PR on Monday.

apolukhin:
@anarthal yes, I saw the links before and followed the approach from the links.

@Lastique noted that there's something wrong with such approach. Including all the standard headers before the `export module` seems to be an ad-hoc solution that scales poorly.

I'm planing to do thefollowin experiment with module fragments: In each header a new module fragment is declared and in PMI all of them are exported

pdimov:
> yes, I saw the links before and followed the approach from the links.

You also need to read the third post, which describes the approach we ultimately settled on.

anarthal:
If we want to support dual builds, I strongly recommend being able to run the library's test suite under modular builds. Otherwise, we're just shipping untested code, which I don't think we should do.

See how the two PRs mentioned in the articles have full test suites and CI builds:

https://github.com/boostorg/mp11/pull/104
https://github.com/boostorg/charconv/pull/255

apolukhin:
> You also need to read the third post, which describes the approach we ultimately settled on.

Oops, I was sure that saw it but now I realize the missed it. Thanks!
I see that the settled approach with `export namespace` is quite close to the libstdc++ approach for `std` [module](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/src/c%2B%2B23/std.cc.in)

@pdimov @anarthal I'll concentrate on  C++20 modularization of libraries I maintain. Please, confirm that I understood the core principles correctly:
* module should be in `module/` directory
* includes should transparently for the user change into `import` depending on `BOOST_USE_MODULES` macro
* module should place all the library entities into a global module fragment and use `export namespace`. 
* module build requires usable module `std`
* all the tests should also run on the library module

anarthal:
I'd say so, yes. Also, remember to add the relevant CMake and CI code. I'm happy to review the PRs for your libraries if it helps.

PR #203, opened 2025-09-08
detail::span_convertible handles void types

### LINKED ISSUE #202: detail::span_is_range is not SFINAE-friendly
Calling span with a single element that partially matches with `span(R&& r)` leads to a compile error: https://godbolt.org/z/hzfcdd8f6

https://github.com/boostorg/core/blob/c6b098d25d32ea5f53950492c19904143bf3e6f1/include/boost/core/span.hpp#L227

ashtum:
fixes #202

### DIFF
--- a/include/boost/core/span.hpp
+++ b/include/boost/core/span.hpp
@@ -23,9 +23,17 @@ class span;
 
 namespace detail {
 
+template<class U, class T, class = void>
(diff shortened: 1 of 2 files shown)


ashtum:
@pdimov , any chance this could be merged for this release?

PR #205, opened 2025-10-09
Use the do {} while (false) idiom for BOOST_TEST_THROWS() and BOOST_TEST_NO_THROW()

gennaroprota:
Reason: Common hygiene for multi-line macros. Makes things like

```
if (condition)
    BOOST_TEST_THROWS(...);
else
    ...
```

work correctly. Also, it doesn't generate an empty statement (which compilers may warn about) when the user adds a semicolon (which they usually do).


pdimov:
@Lastique @glenfe Any objections to merging this?

pdimov:
Looks like this is the only such occurrence though.

Issue #206, opened 2025-10-16
Overload resolution ambiguity between std::data and boost::data on boost::span


DiegoBaldassarMilleuno:
The following code fails to compile (in C++17 and up) because of an ambiguous resolution between `std::data` and `boost::data` (ADL-found):

```c++
#include <boost/core/span.hpp>
#include <iterator>

int* f(boost::span<int> spn) {
	using std::data;
	return data(spn);
}
```

<details>
  <summary>Output of GCC15:</summary>

```
./test.cpp: In function 'int* f(boost::span<int>)':
./test.cpp:19:20: error: call of overloaded 'data(boost::span<int>&)' is ambiguous
   19 |         return data(spn);
      |                ~~~~^~~~~
./test.cpp:19:20: note: there are 8 candidates
In file included from /opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/array:46,
                 from boost_1_89_0/boost/core/span.hpp:13,
                 from ./test.cpp:1:
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/range_access.h:335:5: note: candidate 1: 'constexpr decltype (__cont.data()) std::data(const _Container&) [with _Container = boost::span<int>; decltype (__cont.data()) = int*]'
  335 |     data(const _Container& __cont) noexcept(noexcept(__cont.data()))
      |     ^~~~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/range_access.h:324:5: note: candidate 2: 'constexpr decltype (__cont.data()) std::data(_Container&) [with _Container = boost::span<int>; decltype (__cont.data()) = int*]'
  324 |     data(_Container& __cont) noexcept(noexcept(__cont.data()))
      |     ^~~~
In file included from boost_1_89_0/boost/core/span.hpp:12:
boost_1_89_0/boost/core/data.hpp:18:1: note: candidate 3: 'constexpr decltype (c.data()) boost::data(C&) [with C =span<int>; decltype (c.data()) = int*]'
   18 | data(C& c) noexcept(noexcept(c.data())) -> decltype(c.data())
      | ^~~~
boost_1_89_0/boost/core/data.hpp:25:1: note: candidate 4: 'constexpr decltype (c.data()) boost::data(const C&) [with C = span<int>; decltype (c.data()) = int*]'
   25 | data(const C& c) noexcept(noexcept(c.data())) -> decltype(c.data())
      | ^~~~
```

</details>

[...]

PR #207, opened 2025-10-16
Use `std::data` and `std::size` for boost equivalents, when possible

Lastique:
This avoids the potential ambiguity between `boost::` and `std::` functions when the user calls `data()` or `size()` unqualified and both alternatives are found via ADL.

Fixes https://github.com/boostorg/core/issues/206.


pdimov:
I like having a test that checks exactly what's being reported as a problem; in this case, invoking `data(sp)` unqualified on a `boost::span` when `using std::data;` is in effect.

Lastique:
Any objections to merging this?

PR #209, opened 2025-12-08
Fix boostorg/core#208: Improve null_deleter documentation

### LINKED ISSUE #208: null_deleter docs could use improvement
The null_deleter docs could also explain that it it useful when you don't have a dynamically allocated object at all. For example
```
struct impl { ... };

shared_ptr<impl> get_default_impl()
{
  static impl def;
  return shared_ptr<impl>( &def, null_deleter );
}
```

vinniefalco:
Add example showing use of null_deleter with static objects. The documentation now includes two examples:
1. Using null_deleter with external objects (e.g., std::cout)
2. Using null_deleter with static objects

Closes #208

### DIFF
--- a/doc/null_deleter.qbk
+++ b/doc/null_deleter.qbk
@@ -20,13 +20,39 @@ which can be used as a deleter with smart pointers such as `unique_ptr` or `shar
 deleter doesn't do anything with the pointer provided upon deallocation, which makes it useful
 when the pointed object doesn't need to be deallocated or is deallocated elsewhere.
 
-[section Example]
+[section Examples]
+
+[section Using null_deleter with external objects]
+
+The `null_deleter` is useful when wrapping a pointer to an external object (such as `std::cout`)
+that should not be deallocated when the smart pointer goes out of scope.
+
 ``
 std::shared_ptr< std::ostream > make_stream()
 {
     return std::shared_ptr< std::ostream >(&std::cout, boost::null_deleter());
 }
 ``
+
+[endsect]
+
+[section Using null_deleter with static objects]
+
+The `null_deleter` is also useful when you need to return a smart pointer to a static object.
+This allows you to return a `shared_ptr` to a static instance without attempting to delete it.
+
+``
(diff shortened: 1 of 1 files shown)


vinniefalco:
It may seem obvious to you that the two examples are substantially identical yet for a beginner, seeing the variable defined explicitly and then used with the aliasing constructor may provide a level of clarity that you already enjoy thanks to your considerable experience.

That said, there is even room for a third example. One that demonstrates that `null_deleter` is not even necessary in C++11 as one could simply pass an empty lambda: `[](T*){ }` and avoid the need to include anything at all.

Lastique:
> It may seem obvious to you that the two examples are substantially identical yet for a beginner, seeing the variable defined explicitly and then used with the aliasing constructor may provide a level of clarity that you already enjoy thanks to your considerable experience.

I don't see what confusion the variable declaration is supposed to clarify.

> That said, there is even room for a third example. One that demonstrates that `null_deleter` is not even necessary in C++11 as one could simply pass an empty lambda: `[](T*){ }` and avoid the need to include anything at all.

First, that example would be irrelevant to `null_deleter` as it would not illustrate its usage. Second, that example would not be applicable when the type of the deleter need to be specified explicitly, e.g. in `std::unique_ptr`.

I don't mind having more examples in the docs, but the examples have to illustrate different usages of the tool, or its different features. Having multiple essentially the same examples is not improving the documentation, it is adding bloat which deters the reader by the sheer amount of reading. The documentation should be succinct yet informative.

You could replace the existing example with a better one to illustrate its usage. However, I don't think your proposed example is better either as it contains underspecified elements such as the `impl` class, is more abstract, and at the same time it is larger than the original one. So I think, keeping the current example is best, as it shows the real use case of the tool (notably, this was the actual reason why I wrote the deleter in Boost.Log).

Thanks for the issue and the PR. I did add a note in the text that the deleter can be used when the object need not be deleter. That was already of implied by the rest of the text, but maybe mentioning this explicitly is better.

